PR #6, opened 2021-12-05
Add docc support for SymbolKit

Kyle-Ye:


Kyle-Ye:
Is there anyone in Apple has the the source diagram file of these images and can we export a dark one? (eg. conforms~dark.png)

Or since the diagram is so easy, can I update a new pair(light and dark) of the 3 image to replace them? (Using open-source draw.io to draw which supports light and dark theme)

It now does not look very harmonious with the dark background.

Screen Shot 2021-12-05 at 22 01 39

Kyle-Ye:
@swift-ci please test

QuietMisdreavus (approved):
This looks good, thanks so much! I asked around about the images; we may want to land those separately so that we can have the documentation catalog support ready sooner.

PR #8, opened 2021-12-11
Split Symbol.swift to reduce complexity of the file

Kyle-Ye:
The Symbol.swift currently has 1400 lines.

This PR is intended to split the Symbol.swift file into a file system hierarchy.

So that the Symbol.swift only have SymbolGraph.Symbol definition and also makes it easier for writing test case.

## Discussion
- [x] Which style do you prefer?
```
- Symbol
	- Availability
		- Availability.swift
		- Domain.swift
	- Symbol.swift  	
```

Or

```
- Symbol
	- Availability
		- Domain.swift
	- Availability.swift
- Symbol.swift  	
```

- [x] What should we do if there are 2 struct/enum both named as "Kind"

Currently there are `SymbolGraph.Symbol.DeclarationFragments.Fragment.Kind` and `SymbolGraph.Symbol.Kind`.

Although they are not in the same folder, they are in the same module. Which is not allowed by Swift

```
error: filename "Kind.swift" used twice: 'Symbol/DeclarationFragments/Fragment/Kind.swift' and 'SymbolKit/SymbolGraph/Symbol/Kind.swift'
note: filenames are used to distinguish private declarations with the same name
```

> https://belkadan.com/blog/2021/11/Swift-Mangling-Regret-Private-Discriminators/

## Status
- [x] ~WIP~
After the above 2 discussion is solved, I'll complete the full PR.

PR #9, opened 2021-12-14
update images in docs

QuietMisdreavus:
This PR updates the images used in the README and the docs to allow for the doc catalog added in https://github.com/apple/swift-docc-symbolkit/pull/6 to have proper dark-mode images.

PR #11, opened 2021-12-18
[5.6] Cherry pick "Split Symbol.swift to reduce complexity of the file"

Kyle-Ye:
The Symbol.swift currently has 1400 lines.

This PR is intended to split the Symbol.swift file into a file system hierarchy.

So that the Symbol.swift only have SymbolGraph.Symbol definition and also makes it easier for writing test case.

> Cherry picked from #8 

Since the other 2 diff commit only changed the docs, I think they are low-risk too.

@ethan-kusters Could you please help confirm this?


QuietMisdreavus:
cc @sara-radi, who is the interim 5.6 branch manager while @franklinsch is on vacation.

This looks like it's porting the whole of `main` over? That means it also includes #6 and #9 as well, not just #8. The others do seem like low-risk changes, but i was under the assumption that they weren't meant to be pushed ahead like this. I'll let Sara make the call, though.

sara-radi (approved):

PR #12, opened 2022-01-11
[DNM] Seeking feedback for `SemanticVersion` implementation

WowbaggersLiquidLunch:
The implementation is largely borrowed from SwiftPM's and TSC's `Version` (at least as of the penultimate commit in this PR), but with some changes that make it fully compliant to SemVer 2.0.0. I intend to apply the roughly same changes (the parts that make sense to each) to SwiftPM's, TSC's, and docc's too.

It's not fully implemented yet (test cases are outdated (and should be ignored for now) and lots of documentation is missing), but the logic is basically done. Because of the large number of files renamed in the "main" branch, it's difficult to rebase my branch, so even though it's not finished, I'm putting it up here for some feedback.

I estimate that it will probably take another month (but likely sooner) before I get everything ready.


Sources/SymbolKit/SymbolGraph/Misc/SemanticVersionPrerelease.swift
```diff
@@ -0,0 +1,135 @@
...
+		internal enum Identifier {
+			/// A numeric pre-release identifier.
+			/// - Parameter identifier: The identifier.
+			case numeric(_ identifier: Int)
+			/// An alphanumeric pre-release identifier.
+			/// - Parameter identifier: The identifier.
+			case alphanumeric(_ identifier: String)
+		}
+	}
+}
```
WowbaggersLiquidLunch, line 26: Because the precedence between pre-release identifiers depends on whether they're numeric, their numeric-ness is recorded this way right after each identifier is validated during the version's initialization.

WowbaggersLiquidLunch:
closing this because it's superseded by #36

PR #21, opened 2022-04-04
add `ivar` and `macro` to the known symbol kinds

QuietMisdreavus:
Bug/issue #, if applicable: rdar://91166981

## Summary

To allow SymbolKit to load symbol graphs from the in-progress Clang symbol-graph support (cf. https://reviews.llvm.org/D122611 and https://reviews.llvm.org/D122446), two new symbol kinds need to be added: One for C preprocessor macros, and one for Objective-C interface variables. This PR also adds a new test that ensures that all the `KindIdentifier` cases can parse back as themselves, to ensure that the `.init(identifier:)` and `lookupIdentifier()` functions don't get out of sync.

## Dependencies

None

## Testing

The "ExtractAPI" symbol graph support in Clang is still experimental, but symbol graphs can be found as part of [the ExtractAPI tests](https://github.com/llvm/llvm-project/tree/main/clang/test/ExtractAPI) in the LLVM main branch. Loading either the `macros.c` output or the `objc_interface.m` output should result in a parsed `SymbolGraph` that has no `.unknown` symbols.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -23,6 +23,8 @@ extension SymbolGraph.Symbol {
         case `func`
         case `operator`
         case `init`
+        case ivar
+        case macro
         case method
         case property
         case `protocol`
@@ -53,6 +55,8 @@ extension SymbolGraph.Symbol {
             case .func: return "func"
             case .operator: return "func.op"
             case .`init`: return "init"
+            case .ivar: return "ivar"
+            case .macro: return "macro"
             case .method: return "method"
             case .property: return "property"
             case .protocol: return "protocol"
@@ -86,6 +90,8 @@ extension SymbolGraph.Symbol {
             case "func": return .func
             case "func.op": return .operator
             case "init": return .`init`
+            case "ivar": return .ivar
+            case "macro": return .macro
             case "method": return .method
             case "property": return .property
             case "protocol": return .protocol
--- a/Tests/SymbolKitTests/SymbolGraph/Symbol/SymbolKindTests.swift
+++ b/Tests/SymbolKitTests/SymbolGraph/Symbol/SymbolKindTests.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

daniel-grumberg (approved):
LGTM, although we should discuss behavior for function like macros. This is not something we currently support in clang, but we really should.

franklinsch (approved):
LGTM. Is there additional work required here for these symbols to show up in DocC?

QuietMisdreavus:
@franklinsch As far as i know, the only work needed is on the Clang side itself - so far they've kept within the primary fields that SymbolKit parses. In the future they may want to add some mixins of their own (e.g. template parameters), which we'll either need to add directly or handle via a "generic mixin" type. For now, though, they should be able to be loaded no problem.

@daniel-grumberg Are you thinking about using `functionSignature` to encode the macro parameters? It looks like [the `FunctionParameter` type](https://github.com/apple/swift-docc-symbolkit/blob/main/Sources/SymbolKit/SymbolGraph/Symbol/FunctionSignature/FunctionParameter.swift) only needs a name, and the declaration fragments can be whatever you need. Any additional behavior on top of that would need to be handled by Swift-DocC; all SymbolKit cares about is that the data is well-formed.

Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
```diff
@@ -23,6 +23,8 @@ extension SymbolGraph.Symbol {
         case `func`
         case `operator`
         case `init`
+        case ivar
```
Kyle-Ye, line 26: Nit:
Update the year in this file too
```
Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
```

Kyle-Ye (approved):

Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
QuietMisdreavus: Good catch! Done.

QuietMisdreavus:
@swift-ci Please test

daniel-grumberg:
> @franklinsch As far as i know, the only work needed is on the Clang side itself - so far they've kept within the primary fields that SymbolKit parses. In the future they may want to add some mixins of their own (e.g. template parameters), which we'll either need to add directly or handle via a "generic mixin" type. For now, though, they should be able to be loaded no problem.
> 
> @daniel-grumberg Are you thinking about using `functionSignature` to encode the macro parameters? It looks like [the `FunctionParameter` type](https://github.com/apple/swift-docc-symbolkit/blob/main/Sources/SymbolKit/SymbolGraph/Symbol/FunctionSignature/FunctionParameter.swift) only needs a name, and the declaration fragments can be whatever you need. Any additional behavior on top of that would need to be handled by Swift-DocC; all SymbolKit cares about is that the data is well-formed.

Currently we don't emit parameter information etc for macros because we don't have a good way of grabbing code comments. It would also be a bit awkward because macro parameters are untyped. We are going to need something down the line for C++ templates, although I would have thought we do something for generic parameters in Swift already, how would it work currently for:
```swift
func swapTwoValues<T>(_ a: inout T, _ b: inout T) {
    let temporaryA = a
    a = b
    b = temporaryA
}
```

QuietMisdreavus:
@daniel-grumberg The `functionSignature` type doesn't need to have types for parameters, as far as i know. Simply having the name is sufficient. Individual function parameters also don't receive their own doc comments - in swift-docc, the preference is to have a list of "Parameters" in the function's doc comment, which can then be matched against the function signature to double-check that the actual function parameters are being documented.

As for the Swift sample, there is currently a `swiftGenerics` mixin that describes type parameters and constraints. Since the name is language-specific, i imagine C++ templates would need to introduce their own mixin for documenting those.

PR #22, opened 2022-04-05
Enable library evolution via a conditional

franklinsch:
Bug/issue #, if applicable: rdar://91237393

## Summary

Adds support for building the `SymbolKit` library with library evolution enabled. This is disabled by default, and to enable the behavior, add an `.enable-library-evolution` file at the root of your checkout.

## Dependencies

None.

## Testing

Verify that `SymbolKit` is built with library evolution enabled by inspecting build logs if and only if the checkout has an `.enable-library-evolution` file at its root.

## Checklist

- ~[ ] Added tests~ N/A

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,6 +12,7 @@
 */
 
 import PackageDescription
+import Foundation
 
 let package = Package(
     name: "SymbolKit",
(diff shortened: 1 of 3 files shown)


franklinsch:
@swift-ci please test

Package.swift
```diff
@@ -23,9 +24,29 @@ let package = Package(
...
+func librarySwiftSettings() -> [SwiftSetting]? {
+    let manifestLocation = URL(fileURLWithPath: #filePath)
+
+    let enableLibraryEvolutionFileLocation = manifestLocation
+        .deletingLastPathComponent()
+        .appendingPathComponent(".enable-library-evolution")
+    
+    // If there is a `.enable-library-evolution` file as a sibling of this package manifest
+    // build libraries with library evolution enabled.
+    if FileManager.default.fileExists(atPath: enableLibraryEvolutionFileLocation.path) {
```
Kyle-Ye, line 45: Maybe we should add the logic to Package@swift-5.5.swift as well since most of the clients are using swift 5.5+.

franklinsch: Yes! Thanks for pointing that out. Fixed

Kyle-Ye:
Also, after adding the `.enable-library-evolution` file in the root and then building it, I 
still could not find any evidence it is using library evolution mode. (No relevant flag in the build log nor the xx.moduleinterface file)

franklinsch:
@Kyle-Ye Hmm, I am not able to reproduce this, I can see `-enable-library-evolution` on my end. To be clear, you checked out this branch and ran `touch .enable-library-evolution` at the root of the checkout?

Do you see the flag when you run `swift build -v` from the command line?

Kyle-Ye (approved):

franklinsch:
This change didn't end up suiting my needs because I couldn't find a reliable way to always generate a `.swiftinterface` (e.g., as @Kyle-Ye pointed out!), so I will close this PR for now.

PR #24, opened 2022-04-06
Revert "add `ivar` and `macro` to the known symbol kinds"

QuietMisdreavus:
Reverts apple/swift-docc-symbolkit#21

Adding new KindIdentifier cases apparently caused swift-docc to stop compiling, which is breaking CI in Swift. I'm reverting this for now so i can land it alongside the required DocC change.

PR #26, opened 2022-04-06
Revert "Revert "add `ivar` and `macro` to the known symbol kinds" (#24)"

QuietMisdreavus:
This reverts commit 3cd58e73390d9d1e6547e2b2cc544cc867d8718f.

Bug/issue #, if applicable: rdar://91166981

This PR re-applies the changes from #21. A corresponding change in Swift-DocC (https://github.com/apple/swift-docc/pull/126) is required to land quickly afterward.

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -23,6 +23,8 @@ extension SymbolGraph.Symbol {
         case `func`
         case `operator`
         case `init`
+        case ivar
+        case macro
         case method
         case property
         case `protocol`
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

Kyle-Ye (approved):

Sources/SymbolKit/SymbolGraph/Symbol/KindIdentifier.swift
```diff
@@ -23,6 +23,8 @@ extension SymbolGraph.Symbol {
         case `func`
         case `operator`
         case `init`
+        case ivar
```
ethan-kusters, line 26: Should we consider converting these two statically defined properties to avoid API breakage like this in the future?

It doesn't seem like we should be promising a totally stable API here via enum cases.

It will be API breaking either way since clients will need to add a new `default` case but at least we'll avoid breaking API in the future.

daniel-grumberg: +1 here. I think we need to add a default unknown kind to avoid these sorts of breakages.

QuietMisdreavus: I like this suggestion! I imagine we'll want to add more symbol kinds in the future, and implying that consumers can exhaustively match them is perhaps a little misleading, if we want to properly support _any_ symbol kind that a potential symbol graph could offer (cf. #7).

Could we change Swift-DocC to use `case default` branches first, though? It should still work with an enum, and the DocC code defaults to an `.unknown` symbol kind anyway. Then we could land this type definition change to `KindIdentifier` here without breaking DocC or toolchain builds.

QuietMisdreavus:
Based on @ethan-kusters's comment and offline discussion, i'm closing this series of PRs in favor of an alternate implementation path that should avoid breaking Swift CI.

PR #29, opened 2022-04-14
add tool to dump a unified symbol graph from given standalone graphs

QuietMisdreavus:
Bug/issue #, if applicable: n/a

## Summary

This PR adds `Encodable` conformances to several unified symbol graph types, and adds a new executable target to load several symbol graphs, merge them into a unified symbol graph, and dump the result to a file or to standard output. This introduces a dependency on swift-argument-parser, so i added the version that swift-docc was using, for consistency's sake.

I mainly wanted to add this so i could debug the unified graph behavior when matching symbols together.

## Dependencies

n/a

## Testing

This debug tool can be run with either a list of individual symbol graphs given as positional command-line arguments, or as a directory which is then recursively scanned, to mirror swift-docc's behavior. By default, it will dump pretty-printed JSON to standard out, but this can be overridden by passing `--no-pretty-print` or `-o filename`, respectively.

For example, with a directory of symbol graphs:

```
swift run dump-unified-graph --symbol-graph-dir /path/to/symbol-graphs -o MyModule.unified.json
```

By default, if there is only one module represented in the symbol graphs being loaded, the tool will use that module. However, if there is more than one module being represented, the module must be chosen with the `--module-name` flag. For example:

```
swift run dump-unified-graph --symbol-graph-dir /path/to/symbol-graphs --module-name MyModule -o MyModule.unified.json
```

## Checklist

- [ n/a ] Added tests


QuietMisdreavus:
@swift-ci Please test

Package.swift
```diff
@@ -27,5 +31,26 @@ let package = Package(
...
+                "SymbolKit",
+                .product(name: "ArgumentParser", package: "swift-argument-parser"),
+            ]),
     ]
 )
+
+// If the `SWIFTCI_USE_LOCAL_DEPS` environment variable is set,
+// we're building in the Swift.org CI system alongside other projects in the Swift toolchain and
+// we can depend on local versions of our dependencies instead of fetching them remotely.
+if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
```
ethan-kusters, line 46: Should we just exclude the `dump-unified-graph` target when building in SwiftCI?

Package.swift
```diff
@@ -12,13 +12,17 @@
...
 import PackageDescription
+import class Foundation.ProcessInfo
 
 let package = Package(
     name: "SymbolKit",
     products: [
         .library(
             name: "SymbolKit",
             targets: ["SymbolKit"]),
+        .executable(
```
ethan-kusters, line 23: Maybe we should start with keeping `dump-unified-graph` as an internal target and not vending it as a product? Since it's mainly useful as a debugging tool within the repo for now.

Sources/dump-unified-graph/main.swift
```diff
@@ -0,0 +1,110 @@
...
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import ArgumentParser
+import SymbolKit
+
+struct DumpUnifiedGraph: ParsableCommand {
```
ethan-kusters, line 15: Should we `@main` this and rename the file to `DumpUnifiedGraph`?

```suggestion
@main
struct DumpUnifiedGraph: ParsableCommand {
```

Package.swift
QuietMisdreavus: How do we do that? Does that mean dropping it from the `products` listing? Once that's done, would i rework the dependencies block to always pull ArgumentParser from GitHub?

ethan-kusters: Ah I think we would essentially keep this block but remove the `else` statement and add:

```swift
package.targets += [
    .executableTarget(
         name: "dump-unified-graph",
         dependencies: [
             "SymbolKit",
             .product(name: "ArgumentParser", package: "swift-argument-parser"),
         ]),
]
```

Does that make sense? It's definitely a _little_ odd but I think worth it to reduce risk in SwiftCI.

QuietMisdreavus: After looking at it a little bit, i thought it would look like that. I've pushed up a couple commits to address your comments. ✌️

QuietMisdreavus:
@swift-ci Please test

Package.swift
```diff
@@ -29,3 +30,22 @@ let package = Package(
...
+// we're building in the Swift.org CI system alongside other projects in the Swift toolchain.
+// In this case we only want to build the `dump-unified-graph` tool outside of Swift CI.
+if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
+    // Building standalone, so fetch all dependencies remotely.
+    package.dependencies += [
+        .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.0.1")),
+    ]
+
+    package.targets += [
+        .executableTarget(
```
ethan-kusters, line 44: I _think_ `executableTarget` wasn't introduced until around Swift 5.4 so I this is technically incompatible with the 5.2 package description here.

Maybe we should just only add the new target to the `5.5` package description for simplicity? Since the main goal of the the `5.2` one is backwards compatibility with packages that depend on SymbolKit.

Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
```diff
@@ -98,3 +98,159 @@ extension UnifiedSymbolGraph {
         }
     }
 }
+
+extension UnifiedSymbolGraph.Symbol: Encodable {
```
ethan-kusters, line 102: I really like how this is isolated into an extension like this. Maybe it's worth also splitting into a separate file (`UnifiedSymbol+Encodable.swift`)? What do you think?

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import ArgumentParser
+import SymbolKit
+
+@main
+struct DumpUnifiedGraph: ParsableCommand {
+    @Option(help: "module's symbol graph to output (will infer a single module, but will fail if multiple modules are being loaded)")
```
ethan-kusters, line 17: This is just a nit-pick, but I think it's kind of a slick feature of ArgumentParser.

If you replace this with an explicit [`ArgumentHelp(_:discussion:)`](https://apple.github.io/swift-argument-parser/documentation/argumentparser/argumenthelp/init(_:discussion:valuename:visibility:)) initializer you can put the more descriptive text in a discussion section:

```suggestion
    @Option(help: ArgumentHelp("module's symbol graph to output", description: "will infer a single module, but will fail if multiple modules are being loaded")
```

Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
QuietMisdreavus: Should the same be done for `UnifiedSymbolGraph` itself, since its `Encodable` conformance is also in its own extension block?

ethan-kusters: I think that might be nice. Since these models are non-trivial it feels like it's a little easier to parse these different components if we keep them more isolated.

QuietMisdreavus: I can get behind that; since i had to jump through some hoops to write up the format the way i wanted, it's worth keeping it by itself so it doesn't inflate the base file itself. I'll do that.

QuietMisdreavus:
@swift-ci Please test

Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbolGraph+Encodable.swift
```diff
@@ -0,0 +1,46 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2021 Apple Inc. and the Swift project authors
```
ethan-kusters, line 4: Nit:
```suggestion
 Copyright (c) 2022 Apple Inc. and the Swift project authors
```

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+@main
+struct DumpUnifiedGraph: ParsableCommand {
+    @Option(help: ArgumentHelp("module's symbol graph to output", discussion: "will infer a single module, but will fail if multiple modules are being loaded"))
+    var moduleName: String?
+
+    @Flag(inversion: .prefixedNo,
+          help: "whether to pretty-print the output JSON (default: true)")
+    var prettyPrint: Bool = true
+
+    @Option(name: .short, help: "output file to write to (default: standard out)")
```
ethan-kusters, line 24: This looks a little strange/confusing to me in the emitted help menu. Is there a `.shortAndLong` name option maybe?

```
-o <o>                  output file to write to (default: standard out)
```

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+    var output: String?
+
+    @Option(help: "directory to recursively load symbol graphs from", completion: .directory)
```
ethan-kusters, line 27: I didn't know about these `completion` options! We should update `docc` with these.

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import ArgumentParser
+import SymbolKit
+
+@main
+struct DumpUnifiedGraph: ParsableCommand {
```
ethan-kusters, line 16: Is there a general overview we could add to the command via its configuration?

```
static let configuration = CommandConfiguration(abstract: "Does a foo...")
```

ethan-kusters (commented):
This looks good to me! Just had a few other thoughts 😃

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+    @Option(help: ArgumentHelp("module's symbol graph to output", discussion: "will infer a single module, but will fail if multiple modules are being loaded"))
+    var moduleName: String?
+
+    @Flag(inversion: .prefixedNo,
+          help: "whether to pretty-print the output JSON (default: true)")
```
ethan-kusters, line 21: ArgumentParser actually adds the `(default:` text by default (🙃) so it's being emitted twice currently:

```
  --pretty-print/--no-pretty-print
                          whether to pretty-print the output JSON (default: true) (default: true)
```

```suggestion
          help: "whether to pretty-print the output JSON")
```

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+        } else {
+            if unifiedGraphs.count > 1 {
+                print("error: No module was given, but more than one module was represented in the symbol graphs")
+                throw ExitCode.failure
+            } else {
+                outputGraph = unifiedGraphs.values.first!
+            }
+        }
+
+        let encoder = JSONEncoder()
```
ethan-kusters, line 82: I think it's probably a good idea to add the formatting `sortedKeys` option even if they don't request pretty printing so that the output can be consistently diffed.

Sources/dump-unified-graph/DumpUnifiedGraph.swift
```diff
@@ -0,0 +1,109 @@
...
+            print(outString!)
+        }
+    }
+}
+
+func loadSymbolGraphsFromDir(_ dir: String) -> [String] {
+    let enumerator = FileManager.default.enumerator(atPath: dir)
+    var symbolGraphs: [String] = []
+
+    while let filename = enumerator?.nextObject() as? String {
```
ethan-kusters, line 102:
```suggestion
    for case let filename as String in enumerator! {
```

QuietMisdreavus: I'm going off the docs for `FileManager.enumerator()` 🙃 but this looks much nicer, thanks!

QuietMisdreavus: One issue i would have with this is that if `enumerator()` returns nil, this will crash, whereas in the current code it will return no symbol graphs and print an error. The docs don't say _when_ `FileManager.enumerator(atPath:)` will return nil, so i can only guess, but i'd rather guard against it than assume.

QuietMisdreavus:
Just rebased the PR onto the latest `main` and added individual commits for each review comment.

@swift-ci Please test

Sources/dump-unified-graph/DumpUnifiedGraph.swift
ethan-kusters: Yeah when I tried this out locally, I also modified line 127 to:
```swift
guard let enumerator = FileManager.default.enumerator(atPath: dir) else {
    return []
}
```

ethan-kusters (approved):
This looks great! I think this will be a really nice tool for debugging. Thanks for putting it together 😃

PR #33, opened 2022-05-24
drop the Location mixin if it's malformed

QuietMisdreavus:
Bug/issue #, if applicable: rdar://93797795

## Summary

In certain situations, symbol graphs can be generated with a `location` mixin which is missing its `position` field. Currently, this causes SymbolKit to crash. This PR tweaks the logic to instead drop the field if it fails to decode.

## Dependencies

None

## Testing

The following symbol graph contains symbols whose `location` mixin contains a `uri` but no `position`. This should be able to be loaded by SymbolKit without crashing after this change.

[DemoFramework.symbols.json.txt](https://github.com/apple/swift-docc-symbolkit/files/8758933/DemoFramework.symbols.json.txt)

## Checklist

- [ n/a ] Added tests

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Symbol/Symbol.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/Symbol.swift
@@ -187,7 +187,7 @@ extension SymbolGraph {
             case Availability.mixinKey:
                 return try container.decode(Availability.self, forKey: key)
             case Location.mixinKey:
-                return try container.decode(Location.self, forKey: key)
+                return try? container.decode(Location.self, forKey: key)
             case Mutability.mixinKey:
                 return try container.decode(Mutability.self, forKey: key)
             case FunctionSignature.mixinKey:


QuietMisdreavus:
@swift-ci Please test

Sources/SymbolKit/SymbolGraph/Symbol/Symbol.swift
```diff
@@ -187,7 +187,7 @@ extension SymbolGraph {
             case Availability.mixinKey:
                 return try container.decode(Availability.self, forKey: key)
             case Location.mixinKey:
-                return try container.decode(Location.self, forKey: key)
+                return try? container.decode(Location.self, forKey: key)
```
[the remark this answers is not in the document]
QuietMisdreavus: Done! Just pushed a new commit with the test.

franklinsch (approved):

Sources/SymbolKit/SymbolGraph/Symbol/Symbol.swift
ethan-kusters, line 190: Should we consider just making the position property in Location optional?

franklinsch: @QuietMisdreavus can you provide more details regarding what the bug in Swift we’re working around is?

Essentially it’s generating invalid empty Location objects, not missing objects. I believe Location is already optional due to the nature of how mixing get processed.

QuietMisdreavus: Basically, the spot in the symbol-graph generator in the Swift compiler that emits the Location mixin has [a special branch for nodes that came from Clang](https://github.com/apple/swift/blob/82717185c1a553537253ff84e2e561e3e4688f6e/lib/SymbolGraphGen/Symbol.cpp#L413-L432). The `IncludeClangDocs` option that gates that branch is supposed to only be set when symbol graphs are being generated for SourceKit. However, the compiler doesn't explicitly set this when generating symbol graphs at build-time; instead it's just left uninitialized. This isn't normally a problem, but we saw a situation where that is actually being treated as `true` instead of `false`, causing this branch to emit its location data.

However, the `location` field it's adding is invalid, according to SymbolKit: it's missing the `position` information. (See the test i added for an example of how this looks.) Since these symbols ordinarily don't have the Location mixin at all, we decided to discard it instead of salvaging it with potentially-invalid position information.

...also, if we made the `position` property optional, we'd need to update Swift-DocC to account for it, which would expand the scope of the change larger than i'm comfortable with. `>_>`

ethan-kusters: Okay I agree this is the right targeted fix then. Longer term it doesn't seem unreasonable for a client to provide a Location with a URI but no specific `position` so maybe we should consider making that property optional when we decode instead of dropping the URI information entirely.

But since that's a breaking change it seems to risky for a change that's targeting release/5.7.

Kyle-Ye: > Okay I agree this is the right targeted fix then. Longer term it doesn't seem unreasonable for a client to provide a Location with a URI but no specific `position` so maybe we should consider making that property optional when we decode instead of dropping the URI information entirely.
> 
> But since that's a breaking change it seems to risky for a change that's targeting release/5.7.

I think maybe we should create an issue to track the breaking change of Location as a TOOD in case we forgot it? 🤔

QuietMisdreavus: The underlying issue with Location mixins from the Swift compiler was fixed in https://github.com/apple/swift/pull/59037; the issue shouldn't be a problem going forward.

PR #47, opened 2022-07-08
[5.7] don't update a symbol graph's module name if it's an extension

QuietMisdreavus:
**Rationale**: Currently, SymbolKit will overwrite the module name of a symbol graph if it detects that it is an "extension" graph. However, this causes problems in Swift-DocC when the module is actually a cross-import overlay, so this PR stops this behavior so that Swift-DocC has the correct module information.
**Risk**: Low
**Risk Detail**: This is a behavior change for SymbolKit, but a minor one - extension symbol graphs are still unified with the symbols of their extended module.
**Reward**: Medium
**Reward Details**: Having this information available to Swift-DocC allows it to render symbols from cross-import overlay modules correctly.
**Original PR**: https://github.com/apple/swift-docc-symbolkit/pull/46
**Issue**: rdar://94736726 (note that this commit is not tagged with a radar number because the Swift-DocC change is also required)
**Code Reviewed By**: @franklinsch 
**Testing Details**: See https://github.com/apple/swift-docc/pull/332


QuietMisdreavus:
@swift-ci Please test

franklinsch (approved):

PR #50, opened 2022-07-19
Add Slice to replace Snippet Chunks

bitjammer:
SE-0356 latest changes include parsing "slices" out of snippets.
These are not quite the same as the current "chunks", which were
a speculation that eventually formed into "slices". Deprecate chunks
and add the slice data model.

Move snippet presentation code to a line-based format: slices are represented
simply as a name and index range to pull their code lines.

Add the `isVirtual` property to modules, for modules that are created
implicitly to hold relationships. When snippet symbol graphs are created,
a fake module is created to hold the snippets, since snippets do not come
from any one module. This property simplifies what is basically just a
heuristic in DocC to a simple factual check.

rdar://95220716

Required for related changes:

- https://github.com/apple/swift-docc/pull/338
- https://github.com/apple/swift-docc-plugin/pull/20

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Module.swift
+++ b/Sources/SymbolKit/SymbolGraph/Module.swift
@@ -12,7 +12,7 @@ import Foundation
 
 extension SymbolGraph {
     /// A ``Module-swift.struct``  describes the module from which the symbols were extracted..
-    public struct Module: Codable {
+    public struct Module: Codable, Equatable {
         /// The name of the module.
         public var name: String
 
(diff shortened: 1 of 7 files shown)


bitjammer:
@swift-ci Please test

Sources/SymbolKit/SymbolGraph/Symbol/Symbol.swift
```diff
@@ -54,6 +54,9 @@ extension SymbolGraph {
         /// The in-source documentation comment attached to a symbol.
         public var docComment: LineList?
 
+        /// If true, the symbol was created implicitly and not from source.
+        public var isVirtual: Bool
```
franklinsch, line 58: Is there a benefit to declaring individual symbols as snippets, rather than relying on the module being virtual? Can a virtual module contain non-virtual symbols?

bitjammer: Yes, the benefit is that there are cases where you may want to ask if a module is virtual, and other cases where you may want to ask if a symbol is virtual. Having this recorded in both places allows the question to be answered without consulting the graph's relationships, which can be expensive.

Sources/SymbolKit/SymbolGraph/Module.swift
```diff
@@ -13,6 +13,14 @@ import Foundation
 extension SymbolGraph {
     /// A ``Module-swift.struct``  describes the module from which the symbols were extracted..
     public struct Module: Codable {
+        public enum CodingKeys: CodingKey {
+            case name
+            case bystanders
+            case platform
+            case version
+            case isVirtual
+        }
```
franklinsch, line 22: Do we need this definition? Can we rely on the synthesised `CodingKeys` instead?

franklinsch (commented):
These changes look good to me, but we're missing tests for the custom coding logic. Could we please add some @bitjammer

Sources/SymbolKit/SymbolGraph/Module.swift
bitjammer: I guess not.

bitjammer: Fixed.

bitjammer:
Added some round-trip tests for `Module`.

bitjammer:
@swift-ci Please test

Tests/SymbolKitTests/SymbolGraph/File.swift
```diff
@@ -0,0 +1,81 @@
+/*
```
franklinsch, line 1: did you mean to name this file `ModuleTests.swift`?

franklinsch (approved):
Just a small comment regarding the file name of the test, otherwise looks good to me, thanks!

Tests/SymbolKitTests/SymbolGraph/File.swift
bitjammer: Yes indeed, fixed. I miss naming files as they are created...

bitjammer:
@swift-ci Please test

PR #51, opened 2022-07-28
Add Slice to replace Snippet Chunks (#50)

bitjammer:
**Explanation**: This changes the data model for Snippets to match latest changes to SE-0356 Swift Snippets, where "slices" were added to allow documentation authors to inline named regions of snippets instead of the whole snippet. This replaces the original design of "chunks" which was geared more towards "sub snippets".

**Scope**: This change is necessary to adopt slices in DocC and the DocC Plugin, so snippets are understood when found in Symbol Graphs, which is the data vehicle for snippet information.

**Risk**: This change should be low risk, since this makes changes to an optional mix-in to the JSON data and is only realized using the DocC Plugin and not Xcode.

**Testing**: 
- [x] Included unit tests
- [x] Manual integration tests in Xcode by running the *Build Documentation* command on several projects
  - DocC itself (Swift)
  - https://developer.apple.com/documentation/xcode/slothcreator_building_docc_documentation_in_xcode (Swift)
  - https://github.com/mdiep/MMMarkdown (Objective-C)

**Reviewer**: @franklinsch @ethan-kusters 

-----

SE-0356 latest changes include parsing "slices" out of snippets.
These are not quite the same as the current "chunks", which were
a speculation that eventually formed into "slices". Deprecate chunks
and add the slice data model.

Move snippet presentation code to a line-based format: slices are represented
simply as a name and index range to pull their code lines.

Add the `isVirtual` property to modules, for modules that are created
implicitly to hold relationships. When snippet symbol graphs are created,
a fake module is created to hold the snippets, since snippets do not come
from any one module. This property simplifies what is basically just a
heuristic in DocC to a simple factual check.

Added round-trip decoding tests for `SymbolGraph.Module`.

rdar://95220716

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Module.swift
+++ b/Sources/SymbolKit/SymbolGraph/Module.swift
@@ -12,7 +12,7 @@ import Foundation
 
 extension SymbolGraph {
     /// A ``Module-swift.struct``  describes the module from which the symbols were extracted..
-    public struct Module: Codable {
+    public struct Module: Codable, Equatable {
         /// The name of the module.
         public var name: String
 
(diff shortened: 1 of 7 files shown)


franklinsch:
Thanks @bitjammer! Since this change is targeting the `release/5.7` branch, can you please update the PR description to follow the template from https://forums.swift.org/t/swift-5-7-release-process/56316?

bitjammer:
@swift-ci Please test

franklinsch (approved):
Approving this for `release/5.7` as part of [SE-0356](https://github.com/apple/swift-evolution/blob/main/proposals/0356-swift-snippets.md).

PR #54, opened 2023-03-01
Add models for HTTP endpoints

franklinsch:
Bug/issue #, if applicable: 

rdar://101408429
rdar://101409112
rdar://101408868
rdar://101408667

## Summary

Add initial support for symbol graphs to model HTTP endpoints.

## Dependencies

None.

## Testing

See instructions in https://github.com/apple/swift-docc/pull/495

## Checklist

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #55, opened 2023-03-03
Revert "Add models for HTTP endpoints"

DougGregor:
Reverts apple/swift-docc-symbolkit#54. This is breaking Swift CI on Linux.


DougGregor:
@swift-ci please test

DougGregor:
We don't think we need to do this; holding open until we're sure.

PR #59, opened 2023-04-24
add a "references" relationship kind and referenceLocation mixin

QuietMisdreavus:
Bug/issue #, if applicable: rdar://108470705

## Summary

This PR adds a new relationship kind, `references`, that describes a relationship where one symbol references another one in its implementation. It also introduces a `referenceLocation` mixin for relationships that describes the source range where the reference occurs.

The idea behind adding this information to a symbol graph would be to allow for more semantic analyses to be made on projects, for example to find unused symbols within a project. Including the source range also allows for a sort of "offline language server", by providing a data source for a "jump to definition" and "find references" functionality.

The JSON for these relationships would look like this:

```json
{
  "kind": "references",
  "source": "source",
  "target": "target",
  "referenceLocation": {
    "uri": "file://file.swift",
    "range": {
      "start": {
        "line": 14,
        "character": 0
      },
      "end": {
        "line": 14,
        "character": 6
      }
    }
  }
}
```

## Dependencies

None

## Testing

This PR is being used as a baseline specification for future implementations, so demonstration symbol graphs need to be synthesized from whole cloth and cannot be tested in a straightforward manner.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Relationship/ReferenceLocation.swift
+++ b/Sources/SymbolKit/SymbolGraph/Relationship/ReferenceLocation.swift
@@ -1,0 +1,39 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2023 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
@swift-ci Please test

daniel-grumberg (approved):
LGTM

PR #60, opened 2023-06-23
add 'AlternateDeclarations' mixin

QuietMisdreavus:
Bug/issue #, if applicable: rdar://111227476

## Summary

When a symbol graph is decoded, SymbolKit currently deduplicates symbols with the same precise identifier by inspecting their declaration and discarding whichever symbol has an `async` keyword in its declaration fragments (or the shorter symbol if neither are async). However, this can lead to important information being lost. This PR adds a new mixin, `AlternateDeclarations`, that stores the declaration fragments of any symbol discarded in this "deduplication" process.

## Dependencies

None

## Testing

As this is a change in the internal data model, testing would need to be performed by decoding a symbol graph that contained "duplicate" symbols and re-encoding it to see the new `alternateDeclarations` mixin. An automated test has been added to ensure that the information is loaded correctly.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Symbol/AlternateDeclarations.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/AlternateDeclarations.swift
@@ -46,7 +46,7 @@ extension SymbolGraph.Symbol {
             alternateDeclarations.declarations.append(declaration)
             mixins[AlternateDeclarations.mixinKey] = alternateDeclarations
         } else {
-            mixins[AlternateDeclarations.mixinKey] = AlternateDeclarations.init(declarations: [declaration])
+            mixins[AlternateDeclarations.mixinKey] = AlternateDeclarations(declarations: [declaration])
         }
     }
 
--- a/Sources/SymbolKit/SymbolGraph/Symbol/AlternateDeclarations.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/AlternateDeclarations.swift
@@ -1,0 +1,58 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2023 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+extension SymbolGraph.Symbol {
+    /// A mixin to hold alternate declarations of a symbol.
+    ///
+    /// This mixin is created while a symbol graph is decoded, to hold ``DeclarationFragments-swift.struct``
+    /// for symbols which share the same precise identifier.
+    public struct AlternateDeclarations: Mixin, Codable {
+        public static let mixinKey = "alternateDeclarations"
+
+        /// Alternate declarations for this symbol.
+        public var declarations: [DeclarationFragments]
+
+        public init(declarations: [DeclarationFragments]) {
+            self.declarations = declarations
+        }
+
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.singleValueContainer()
+            declarations = try container.decode([DeclarationFragments].self)
+        }
+
+        public func encode(to encoder: Encoder) throws {
+            var container = encoder.singleValueContainer()
+            try container.encode(declarations)
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SymbolKit/SymbolGraph/Symbol/AlternateDeclarations.swift
```diff
@@ -0,0 +1,58 @@
...
+    public var alternateDeclarations: [DeclarationFragments]? {
+        (mixins[AlternateDeclarations.mixinKey] as? AlternateDeclarations)?.declarations
+    }
+
+    internal mutating func addAlternateDeclaration(_ declaration: DeclarationFragments) {
+        if var alternateDeclarations = mixins[AlternateDeclarations.mixinKey] as? AlternateDeclarations {
+            alternateDeclarations.declarations.append(declaration)
+            mixins[AlternateDeclarations.mixinKey] = alternateDeclarations
+        } else {
+            mixins[AlternateDeclarations.mixinKey] = AlternateDeclarations.init(declarations: [declaration])
```
Kyle-Ye, line 49: nit
```suggestion
            mixins[AlternateDeclarations.mixinKey] = AlternateDeclarations(declarations: [declaration])
```

Sources/SymbolKit/SymbolGraph/SymbolGraph.swift
```diff
@@ -62,13 +62,27 @@ public struct SymbolGraph: Codable {
...
+            return result
         } else if rhs.declarationContainsAsyncKeyword() {
-            return lhs
+            var result = lhs
+            result.addAlternateDeclaration(from: rhs)
+            return result
         } else {
             // It's not expected to ever end up here, but if we do, we return the symbol with the longer name
             // to have consistent results.
-            return lhs.names.title.count < rhs.names.title.count ? rhs : lhs
```
Kyle-Ye, line 71: If `title.count` is the same. Then the result of `_symbolToKeepInCaseOfPreciseIdentifierConflict(s1, s2)` will be different `_symbolToKeepInCaseOfPreciseIdentifierConflict(s2, s1)`.

But according to the comment, we expect a consistent results.

Should we use a more "stable" property here? eg. `names.title.hash`

Tests/SymbolKitTests/SymbolGraph/SymbolGraphTests.swift
Kyle-Ye, line 1: Can we add a test case for `AlternateDeclarations.swift` in `SymbolTests.swift` so that we have better coverage.

Tests/SymbolKitTests/SymbolGraph/SymbolGraphTests.swift
```diff
@@ -63,6 +63,19 @@ class SymbolGraphTests: XCTestCase {
         XCTAssertTrue(declaration.declarationFragments.contains(where: { fragment in
             fragment.kind == .externalParameter && fragment.spelling == "completionHandler"
         }))
+
+        // The async declaration should have been saved as an alternate declaration
+        let alternateDeclarations = try XCTUnwrap(symbol.alternateDeclarations)
+        XCTAssertEqual(alternateDeclarations.count, 1)
+        let alternate = try XCTUnwrap(alternateDeclarations.first)
```
Kyle-Ye, line 70: Nit:
Since we've checked the `count == 1`, maybe we can access it directly.
```suggestion
        let alternate = alternateDeclarations[0]
```

Kyle-Ye (approved):
LGTM

Sources/SymbolKit/SymbolGraph/SymbolGraph.swift
QuietMisdreavus: I wouldn't want to use a hash specifically, but we could check the lexicographical sorting of the identifiers if they're equal. That way it's still at least based on something visible.

Tests/SymbolKitTests/SymbolGraph/SymbolGraphTests.swift
QuietMisdreavus: I'm testing the data type in `SymbolGraphTests.swift`. `SymbolTests.swift` seems to be more based around encoding/decoding symbols, which doesn't make sense here since `AlternateDeclarations` isn't decoded from on-disk JSON data.

daniel-grumberg (approved):

PR #76, opened 2024-05-20
use all selectors, not just "main graph selectors", when collecting overloads-per-selector

QuietMisdreavus:
Bug/issue #, if applicable: rdar://128284906

## Summary

In #72, i introduced a processing step for cross-platform overload groups where those groups were different depending on the symbol graph that they appeared in. However, this implementation had a flaw: When cleaning up relationships to superfluous overload groups, i used the overloads' "main graph selectors" to restrict which lists of relationships to iterate over. This meant that such overload groups that appeared in extension symbol graphs were not properly cleaned up, leading to a crash in Swift-DocC. This PR updates the filter to use a new `allSelectors` property, which allows these extension overloads to be handled correctly.

## Dependencies

None

## Testing

Automated testing verifies the expected behavior.

## Checklist

- [x] Added tests


QuietMisdreavus:
@swift-ci Please test

patshaughnessy (approved):
Great fix - thank you!

PR #81, opened 2024-07-12
add a separate overload-group calculation in the unified graph collector

QuietMisdreavus:
Bug/issue #, if applicable: rdar://131639487

## Summary

This PR updates the unified symbol graph collector to optionally create overload groups separately from the previous iteration, where it combined pre-existing overload groups created on individual symbol graphs. To preserve API compatibility (and use outside of unified symbol graphs), the other mechanisms are preserved, but they suffer from a bug. When overloads are added in disjoint sets, whether from an extension or from alternate platforms, the old overload grouping code failed to combine them together. This updated code ensures that all overload groups reflect all of the available symbols, regardless of their source.

## Dependencies

None

## Testing

A companion PR on Swift-DocC will go more in-depth into testing. This PR contains additional automated tests that demonstrate the new behavior.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Symbol/DeclarationFragments/DeclarationFragments+Simplify.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/DeclarationFragments/DeclarationFragments+Simplify.swift
@@ -28,71 +28,103 @@ fileprivate extension UnifiedSymbolGraph.Symbol {
     }
 }
 
+private func overloadFragments(
+    declarationFragments: [SymbolGraph.Symbol.DeclarationFragments.Fragment]?,
+    subHeading: [SymbolGraph.Symbol.DeclarationFragments.Fragment]?,
+    navigator: [SymbolGraph.Symbol.DeclarationFragments.Fragment]?,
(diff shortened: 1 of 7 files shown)


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
I chatted with @daniel-grumberg and @mayaepps about this PR offline. We found a couple things to change:

1. Make sure that protocol implementations are still not grouped with their original requirement.
2. Parameterize the existing tests to make sure they all work with the new code.

I'll work on these points later today.

mayaepps:
A small late-breaking thought 😄

> To preserve API compatibility (and use outside of unified symbol graphs), the other mechanisms are preserved, but they suffer from a bug.

The old code will still suffer from this bug after this change, right? Is that something we should note somewhere so it's not forgotten?

QuietMisdreavus:
@mayaepps That's definitely worth noting! I've wanted to write up a doc article about the overloads implementation at some point. I think in the meantime i can throw a warning into a doc comment.

QuietMisdreavus:
@swift-ci Please test

daniel-grumberg (approved):
LGTM!

PR #94, opened 2025-02-04
Revert "adds support to correctly identify snippet symbolgraphs (#89)"

tshortli:
https://github.com/swiftlang/swift-docc-symbolkit/pull/89 appears to be causing reproducible test failures during Swift compiler PR testing on macOS.

```
Test Case '-[SwiftDocCTests.RenderNodeTranslatorTests testSnippetToCodeListing]' started.
/Users/ec2-user/jenkins/workspace/swift-PR-macos/branch-main/swift-docc/Tests/SwiftDocCTests/Rendering/RenderNodeTranslatorTests.swift:965: error: -[SwiftDocCTests.RenderNodeTranslatorTests testSnippetToCodeListing] : XCTUnwrap failed: threw error "notFound(doc://org.swift.docc.snippets/documentation/Snippets/Snippets)"
/Users/ec2-user/jenkins/workspace/swift-PR-macos/branch-main/swift-docc/Tests/SwiftDocCTests/Rendering/RenderNodeTranslatorTests.swift:965: error: -[SwiftDocCTests.RenderNodeTranslatorTests testSnippetToCodeListing] : failed: caught error: "notFound(doc://org.swift.docc.snippets/documentation/Snippets/Snippets)"
Test Case '-[SwiftDocCTests.RenderNodeTranslatorTests testSnippetToCodeListing]' failed (0.062 seconds).
```

https://ci.swift.org/job/swift-PR-macos/20059/console

PR #95, opened 2025-02-16
Support snippet symbolgraph

heckj:
## Summary

- expanded documentation on mergeSymbolGraph to describe the forceLoading
parameter.
- updated the logic to use the lack of @ and the isVirtual metadata
to determine if a symbolgraph is the primary graph for that module.
- resolves https://github.com/swiftlang/swift-docc-symbolkit/issues/88

Snippet extension graphs being considered primary is the source of non-deterministic results in previewing documentation that include snippets (https://github.com/swiftlang/swift-docc/issues/1084). For the logic within swift-dock's SymbolGraphLoader, they should be considered extension graphs. Since the snippet symbol graph extractor doesn't generate an @ in the file name of the graph, the logic is amended to look at the isVirtual metadata for the module, which is set to true for snippets.

I considered using the name generated by the snippet extractor, or doing some heuristic on the name the generator metadata, but after staring at the data for a while, the SymbolGraph.module.isVirtual seemed a more exact and lasting means of determining if a given symbol graph is "the main one" or not.

(reset of #89, previously reverted, when smoke tests didn't pass)

## Dependencies

Depends on https://github.com/swiftlang/swift-docc/pull/1160 for the CI/smoke tests to pass (in particular, this PR was merged earlier, and reverted, due to tests in swift-docc repeatedly failing with these updates. The PR in swift-docc adds structure to a test fixture so that better represents snippets as extending another symbolgraph, which resolves these test issues.

## Testing

Added unit tests to verify loading functionality

## Checklist

- [X] Added tests
- [X] Updated documentation if necessary

### DIFF
--- a/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
+++ b/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
@@ -58,11 +58,12 @@ extension GraphCollector {
     ///   - inputGraph: The symbol graph to merge in.
     ///   - url: The file name where the given symbol graph is located. Used to determine whether a symbol graph
     ///     contains primary symbols or extensions.
-    ///   - forceLoading: A Boolean value, that defaults to `false`, that indicates whether the graph is merged
-    ///     even though it is a `primary` symbol graph.
+    ///   - forceLoading: Whether or not to force processing an extension symbol graph.
     ///
-    /// If the symbol graph is the `primary` graph, this method ignore merging the graph if it isn't already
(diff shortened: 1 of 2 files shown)


heckj:
Prior (identical) PR was merged in #89 and reverted in #94 when the swift-docc tests failed. The PR https://github.com/swiftlang/swift-docc/pull/1160 resolves those failing tests.

I've verified (manually) this PR works as intended when the upstream PR in place.

QuietMisdreavus (approved):

QuietMisdreavus:
@swift-ci Please test

PR #98, opened 2025-04-02
allow SemanticVersion prerelease field to be int or string

QuietMisdreavus:
Bug/issue #, if applicable: rdar://147937598

## Summary

It turns out that the Swift symbol graph generator emits an integer field for the `prerelease` part of a version, if it's present. This causes a decoding error in SymbolKit, since it's expecting a string. This PR updates the `SemanticVersion` type to allow decoding JSON with either a string or an integer for the `prerelease` field.

## Dependencies

None

## Testing

This has been covered with automated tests.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/SemanticVersion.swift
+++ b/Sources/SymbolKit/SymbolGraph/SemanticVersion.swift
@@ -8,6 +8,36 @@
  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+fileprivate enum PreReleaseVersion: Decodable {
+    case `int`(Int)
+    case `string`(String)
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.singleValueContainer()
+
+        if let intValue = try? container.decode(Int.self) {
+            self = .int(intValue)
+        } else if let stringValue = try? container.decode(String.self) {
+            self = .string(stringValue)
+        } else {
+            throw DecodingError.typeMismatch(
+                Self.self,
+                .init(
+                    codingPath: decoder.codingPath,
+                    debugDescription: "Expected Int or String"))
+        }
+    }
+
+    var stringValue: String {
+        switch self {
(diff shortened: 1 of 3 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SymbolKit/SymbolGraph/SemanticVersion.swift
```diff
@@ -8,6 +8,36 @@
  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+fileprivate enum PreReleaseVersion: Decodable {
```
patshaughnessy, line 11: Any reason you put this private enum at the top of the file? I would usually push private implementation details like this to the bottom. Does it need to be declared before we can use it in `SemanticVersion.init(from:)?

bin/check-source
```diff
@@ -18,7 +18,7 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 function replace_acceptable_years() {
     # this needs to replace all acceptable forms with 'YEARS'
-    sed -e 's/20[12][78901234]-20[12][8901234]/YEARS/' -e 's/20[12][8901234]/YEARS/'
+    sed -e 's/20[12][789012345]-20[12][89012345]/YEARS/' -e 's/20[12][89012345]/YEARS/'
```
patshaughnessy, line 21: What's this all about?

patshaughnessy (approved):
Elegant use of an enum for decoding the two alternatives  👍

Sources/SymbolKit/SymbolGraph/SemanticVersion.swift
QuietMisdreavus: Old habit from writing C too much. 😅 I can move it below if you'd prefer, it's no big deal.

binamaniar (approved):
looks good to me.

PR #111, opened 2026-04-30
Keep ARM 64 bit based arch as the canonical architecture.

sofiaromorales:
Bug/issue #, if applicable:  174845543

## Summary

When merging symbols with a shared selector but different architectures,
we overwrite the information of the already recorded symbol. This causes
the latest recorded symbol to prevail, making the result dependent on the
processing order of symbol graphs.

Symbol information can differ across architectures (e.g., availability, declaration
fragments), so for consistency, ARM 64-bit (Apple Silicon) is now the canonical
architecture. This change ensures that when a SGF generated by a 64 bit arch is available,
its symbol information prevails regardless of the order in which architectures are processed.
If there's none, a different version of the symbol remains.
## Dependencies

N/A

## Testing

Unit test added to verify the correct behaviour.

For a more complete test, using swift-docc, preview the following DocC catalog:
[ArchTest.docc.zip](https://github.com/user-attachments/files/27250809/ArchTest.docc.zip)

## Checklist

- [X] Added tests
- [ ] Updated documentation if necessary (N/A)

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Platform.swift
+++ b/Sources/SymbolKit/SymbolGraph/Platform.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
--- a/Sources/SymbolKit/SymbolGraph/Platform.swift
+++ b/Sources/SymbolKit/SymbolGraph/Platform.swift
@@ -15,6 +15,9 @@ extension SymbolGraph {
          The name of the architecture that this module targets, such as `x86_64` or `arm64`. If the module doesn't have a specific architecture, this may be undefined.
          */
         public var architecture: String?
+        var isAppleSilicon: Bool {
+            architecture == "aarch64" || architecture == "arm64"
+        }
 
         /**
          The platform vendor from which this module came, such as `apple` or `linux`.
--- a/Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
+++ b/Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
--- a/Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
+++ b/Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
@@ -126,6 +126,13 @@ extension UnifiedSymbolGraph {
             if isMainGraph && !self.mainGraphSelectors.contains(selector) {
                 self.mainGraphSelectors.append(selector)
             }
+            // Preserve Apple Silicon symbol information over other architectures.
+            if let existingPlatform = self.modules[selector]?.platform {
+                let incomingPlatform = module.platform
+                if !incomingPlatform.isAppleSilicon && existingPlatform.isAppleSilicon {
+                    return
+                }
+            }
 
             // Add a new variant to the fields that track it
             self.modules[selector] = module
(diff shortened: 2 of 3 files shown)


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus (approved):
I feel like it might be worth being able to distinguish between "macOS x86_64" versus "macOS AArch64" (where the distinction is more likely to cause issues) as different selectors, but that will probably require extra work in Swift-DocC to account for the new symbol data. This works in the meantime to address some nondeterminism in the unified symbol graph, though. Thanks!

Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
```diff
@@ -126,6 +126,16 @@ extension UnifiedSymbolGraph {
             if isMainGraph && !self.mainGraphSelectors.contains(selector) {
                 self.mainGraphSelectors.append(selector)
             }
+            
+            if
+                let moduleArchitecture = module.platform.architecture,
+                let currentArchitecture = self.modules[selector]?.platform.architecture {
+                // Don't overwrite Apple Silicon symbol information with information from other architectures.
+                if !SymbolGraph.Platform.arm64ArchitectureNames.contains(moduleArchitecture) &&
```
sofiaromorales: Rewrite this to make it a computed property and improve readability.

Sources/SymbolKit/SymbolGraph/Platform.swift
```diff
@@ -15,6 +15,11 @@ extension SymbolGraph {
          The name of the architecture that this module targets, such as `x86_64` or `arm64`. If the module doesn't have a specific architecture, this may be undefined.
          */
         public var architecture: String?
+        // ARM 64-bit names.
+        static let arm64ArchitectureNames: Set = ["aarch64", "arm64"]
+        var isAppleSilicon: Bool {
+            Self.arm64ArchitectureNames.contains(architecture ?? "")
+        }
```
d-ronnqvist: nit: I feel that this would be simpler without the set of two values. That also avoid the need for a fallback value when there architecture is `nil`.

```suggestion
        var isAppleSilicon: Bool {
            architecture == "aarch64" || architecture == "arm64"
        }
```

Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
```diff
@@ -126,6 +126,13 @@ extension UnifiedSymbolGraph {
             if isMainGraph && !self.mainGraphSelectors.contains(selector) {
                 self.mainGraphSelectors.append(selector)
             }
+            // Preserve Apple Silicon symbol information over other architectures.
+            if let existingPlatform = self.modules[selector]?.platform {
+                let incomingPlatform = module.platform
+                if !incomingPlatform.isAppleSilicon && existingPlatform.isAppleSilicon {
+                    return
+                }
+            }
```
d-ronnqvist, line 135: nit: If we only need to check two Boolean conditions we can check them both in the one if statement

```suggestion
            if self.modules[selector]?.platform.isAppleSilicon == true, !module.platform.isAppleSilicon {
                return
            }
```

Tests/SymbolKitTests/UnifiedGraph/UnifiedSymbolTests.swift
```diff
@@ -527,5 +527,66 @@ class UnifiedSymbolTests: XCTestCase {
...
+           ],
+           "names": {
+               "title": "PlayingCard x86_64"
+           },
+           "accessLevel": "open"
+        }
+        """
+
+        let aarch64Symbol = try decoder.decode(SymbolGraph.Symbol.self, from: aarch64SymbolLiteral.data(using: .utf8)!)
+        let x86_64Symbol = try decoder.decode(SymbolGraph.Symbol.self, from: x86_64SymbolLiteral.data(using: .utf8)!)
```
d-ronnqvist: FYI: It's always possible to convert a UTF8 sequence into `Data` but the `.data(using:)` method can't reflect that in its API because it needs to have the same return value for all kinds of encodings. If you instead use `Data(theString.utf8)` you'll get a non-optional value so you don't have to force unwrap anything.

```suggestion
        let aarch64Symbol = try decoder.decode(SymbolGraph.Symbol.self, from: Data(aarch64SymbolLiteral.utf8))
        let x86_64Symbol = try decoder.decode(SymbolGraph.Symbol.self, from: Data(x86_64SymbolLiteral.utf8))
```

Sources/SymbolKit/UnifiedSymbolGraph/UnifiedSymbol.swift
sofiaromorales: It could be shorter as you suggest but I prefer the readability of the current implementation.

Sources/SymbolKit/SymbolGraph/Platform.swift
sofiaromorales: Resolved, thx! 👍

d-ronnqvist (approved):

sofiaromorales:
@swift-ci please test

PR #113, opened 2026-07-23
Deterministically merge relationships in unified graph

snprajwal:
Bug/issue #, if applicable: rdar://183034556

## Summary

`mergeRelationships` returns an `Array(map.values)`, whose ordering is not stable. The deduplication also operates on a first-seen basis, leading to cases where duplicate relationships with different target fallbacks are non-deterministically chosen to be retained.

Deduplicate using the target fallback instead, and use the first-seen order to ensure that the merge output is stable. Also sort the extension graphs lexicographically by path to ensure that the merge input is stable.

## Dependencies

N/A

## Testing

1. Create two symbol graphs with duplicate relationships (same source, target, kind) but different target fallbacks.
2. Attempt to merge the relationships with `mergeRelationships(:)` multiple times.
3. Previously, the output would change across runs. With this patch, it remains consistent.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
+++ b/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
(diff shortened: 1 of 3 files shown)


snprajwal:
@swift-ci test

QuietMisdreavus (approved):

PR #114, opened 2026-08-03
Fix `moduleNameFor` logic for cross-import overlays.

sofiaromorales:
Bug/issue #, if applicable:  rdar://183708988

## Summary

The method `moduleNameFor`[1] returns the name of the module given a Symbol Graph (SG) and the
SG URL, along with whether it's a main SG. Part of the logic of this method detects when the SG is a cross-import overlay [2].

A comment in this method states that when bystanders are present the SG is a cross-import
overlay [3], but that specific piece of logic is only executed if the `graph.module.bystanders`
property is `nil`, so it never applies to overlays.

That condition was introduced deliberately in [5], on the grounds that `_A_B@B.symbols.json` holds
symbols that extend `B`. For cross-import overlays that is wrong. `_A_B@B.symbols.json` contains symbols 
that extend `A`.

DocC did not hit this error until now because it kept its own copy of this logic. That copy was removed in [4].

This code change fixes the logic so if the SG is coming from an overlay, the module name comes from the SG data `module.name`.

This matches the (now removed) DocC implementation of this method [4].

This commit also updates the test so it matches the expected behaviors.

rdar://183708988

---

[1] https://github.com/swiftlang/swift-docc-symbolkit/blob/main/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift#L134
[2] https://forums.swift.org/t/cross-import-overlays/36710
[3] https://github.com/swiftlang/swift-docc-symbolkit/blob/main/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift#L143
[4] https://github.com/swiftlang/swift-docc/commit/b577c7220a6c7f10b2b7c7fcb98457ebf92a3921
[5] https://github.com/swiftlang/swift-docc-symbolkit/commit/3fc0c6880f712ce38137d0752f24b568fce1522e

## Dependencies

N/A

## Testing

N/A

## Checklist

- [X] Added tests
- [X] Updated documentation if necessary

### DIFF
--- a/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
+++ b/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
@@ -139,7 +139,11 @@ extension GraphCollector {
         let isMainSymbolGraph = !url.lastPathComponent.contains("@") && !graph.module.isVirtual
 
         let moduleName: String
-        if isMainSymbolGraph && graph.module.bystanders == nil {
+
+        // FIXME: Stop forcing cross-import overlays into this branch when Swift-DocC corrects its rendering behavior
+        // (https://github.com/swiftlang/swift-docc/issues/1611)
+        let isCrossImportOverlay = graph.module.bystanders != nil
+        if isMainSymbolGraph || isCrossImportOverlay {
             // When bystander modules are present, the symbol graph is a cross-import overlay, and
             // we need to preserve the original module name to properly render it. It is still
             // kept with the extension symbols, due to the merging behavior of UnifiedSymbolGraph.
--- a/Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
+++ b/Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Copyright (c) 2022-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -183,8 +183,9 @@ class GraphCollectorTests: XCTestCase {
         }
         let a_At_B = try JSONDecoder().decode(SymbolGraph.self, from: jsonDataAatB)
 
-        let (extendedB, extendedBIsMain) = GraphCollector.moduleNameFor(a_At_B, at: .init(fileURLWithPath: "_A_B@B.symbols.json"))
-        XCTAssertFalse(extendedBIsMain)
-        XCTAssertEqual("B", extendedB)
+        // For cross-import overlays the name comes from the graph's `module.name`, not from the module named after the "@" in the file name.
+        let (overlayAtBName, overlayAtBIsMain) = GraphCollector.moduleNameFor(a_At_B, at: .init(fileURLWithPath: "_A_B@B.symbols.json"))
+        XCTAssertFalse(overlayAtBIsMain)
+        XCTAssertEqual("A", overlayAtBName)
     }
 }


sofiaromorales:
@swift-ci please test

heckj (approved):

heckj:
from experience, it's also worth kicking off a smoke-test, as there's secondary effects in high level tests that this could impact.

@swift-ci please smoke test

QuietMisdreavus:
> That condition was introduced deliberately in [5], on the grounds that `A_B@B.symbols.json` holds
symbols that extend `B`. For cross-import overlays that is wrong. `_A_B@B.symbols.json` contains symbols
that extend `A`.

This strikes me as strange, and possibly incorrect. Symbols from `_A_B` that extend `A` (which i'm assuming in this case is the declaring module) should still be in the `_A_B@A` symbol graph. Where are you seeing that they're being placed in the `@B` symbol graph instead? That should be a bug raised with the Swift symbol graph extractor.

sofiaromorales:
@swift-ci please test

snprajwal:
> This strikes me as strange, and possibly incorrect. Symbols from `_A_B` that extend `A` (which i'm assuming in this case is the declaring module) should still be in the `_A_B@A` symbol graph. Where are you seeing that they're being placed in the `@B` symbol graph instead? That should be a bug raised with the Swift symbol graph extractor.

I think what they mean to convey is that symbols extending `B` in `_A_B@B` still sit under `A` as the declaring module, and must hence be associated with `A` when being documented.

Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
snprajwal, line 182: This branch is now unreachable as cross-import overlays are handled in the `graph.module.bystanders != nil` block above. Note that it also appears to return `B` in this case, which is incorrect as per the intention of this PR, which expects `A` here.

snprajwal, line 128: It would be good to update this to also document cross-import overlays.

Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
```diff
@@ -139,7 +139,7 @@ extension GraphCollector {
         let isMainSymbolGraph = !url.lastPathComponent.contains("@") && !graph.module.isVirtual
 
         let moduleName: String
-        if isMainSymbolGraph && graph.module.bystanders == nil {
+        if isMainSymbolGraph || graph.module.bystanders != nil {
```
snprajwal: It is currently not clear as to what exactly this condition checks. It might be better to extract the check into a more obviously named variable instead:


```suggestion
        let isCrossImportOverlay = graph.module.bystanders != nil
        if isMainSymbolGraph || isCrossImportOverlay {
```

Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
snprajwal, line 1: This file would require a copyright bump

Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
```diff
@@ -172,7 +172,7 @@ class GraphCollectorTests: XCTestCase {
         }
         let a_At_A = try JSONDecoder().decode(SymbolGraph.self, from: jsonDataAatA)
 
-        let (extendedA, extendedAIsMain) = GraphCollector.moduleNameFor(a_At_A, at: .init(fileURLWithPath: "_A_B@A.symbols.json"))
+        var (extendedA, extendedAIsMain) = GraphCollector.moduleNameFor(a_At_A, at: .init(fileURLWithPath: "_A_B@A.symbols.json"))
```
snprajwal: It's logically clearer to create new named variables below instead of mutating existing ones for storing state.

QuietMisdreavus:
> I think what they mean to convey is that symbols extending `B` in `_A_B@B` still sit under `A` as the declaring module, and must hence be associated with `A` when being documented.

Isn't this the opposite of the logic we use for extensions? The cross-import overlay symbol graphs are named like that so that they would fall into the logic of extensions, where symbols are filtered into the target module's grouping (unless extension block symbols are used, in which case the `A@B` symbol graph is never created). Wouldn't this affect which unified graph a symbol is sorted into?

QuietMisdreavus (changes requested):
Blocking merge until we can resolve the discussion

Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
```diff
@@ -183,8 +183,8 @@ class GraphCollectorTests: XCTestCase {
         }
         let a_At_B = try JSONDecoder().decode(SymbolGraph.self, from: jsonDataAatB)
 
-        let (extendedB, extendedBIsMain) = GraphCollector.moduleNameFor(a_At_B, at: .init(fileURLWithPath: "_A_B@B.symbols.json"))
-        XCTAssertFalse(extendedBIsMain)
-        XCTAssertEqual("B", extendedB)
+        (extendedA, extendedAIsMain) = GraphCollector.moduleNameFor(a_At_B, at: .init(fileURLWithPath: "_A_B@B.symbols.json"))
```
snprajwal: Here, module B is being extended, while A declares the overlay. This can be named in a less misleading manner:


```suggestion
        // `A` declares the overlay, so that is the module name despite `B` being extended
        let (declaringA, declaringAIsMain) = GraphCollector.moduleNameFor(a_At_B, at: .init(fileURLWithPath: "_A_B@B.symbols.json"))
```

snprajwal (dismissed):

sofiaromorales:
This PR brings back the behavior that was introduced in https://github.com/swiftlang/swift-docc/pull/333 and was mistakenly modified in https://github.com/swiftlang/swift-docc-symbolkit/commit/3fc0c6880f712ce38137d0752f24b568fce1522e. We can discuss this further, but it isn't introducing new logic, just fixing an error when pairing the DocC method declaration with the SymbolKit one.

Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
sofiaromorales: Is this a blocking comment?

Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
sofiaromorales: "unreachable" isn't quite right. It's unreachable only for graphs that carry `bystanders` metadata. A file named like an overlay whose graph doesn't declare bystanders would still land here.

I'd prefer to keep the removal out of this PR. The goal here is a minimal fix for the downstream regression caused in DocC (https://github.com/swiftlang/swift-docc/commit/b577c7220a6c7f10b2b7c7fcb98457ebf92a3921). This can be made in a follow up PR IMO.

QuietMisdreavus:
I did some reading to figure out the original context for why the Swift-DocC version was written that way, and realized that *i* was the one that wrote that version! 😅 The context is spelled out in https://github.com/swiftlang/swift-docc/pull/332, but to summarize that summary, we need to specifically break the "extension-style" handling for extensions from cross-import overlays because of how DocC renders the final pages, which includes module name references for the symbol-graph module (i.e. the module of the unified graph it came from), plus any relevant "bystander modules" if the symbol is from a cross-import overlay. However, because the symbol's module is the same as the bystander module (in the `_A_B@B` scenario), it prints `B` twice in the final page.

I... don't like this as an API design, but as a quick bug-fix to restore the existing behavior i can be okay with it. I think the ultimate fix would be for Swift-DocC to figure out how to pull out the declaring module and use that when printing the modules into the rendered page, but that's a bigger task. We should either land this or revert the PR in DocC that removed its copy of the function.

sofiaromorales:
Thanks for the research @QuietMisdreavus!

I'd rather land this than revert. Reverting brings back a local DocC copy of a SymbolKit method, fixing it here keeps a single implementation and puts the behavior in what is now the _de facto_ method. Any other caller of this API would otherwise get the wrong module name in this scenario.

QuietMisdreavus:
> Any other caller of this API would otherwise get the wrong module name in this scenario.

Landing the PR (or reverting the other) would cause SymbolKit to return the "wrong" module name. The thing we're fixing is the work-around in DocC to allow it to print the correct information. The correct fix is to change that behavior, but in the meantime there's a regression that we need to patch.

Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
QuietMisdreavus: I would vastly prefer that we keep the test using new variables here rather than overwriting the previous ones. We don't need to actually modify the data, and declaring it as a mutable variable creates unnecessary confusion.

Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
QuietMisdreavus:
```suggestion
        // FIXME: Stop forcing cross-import overlays into this branch when Swift-DocC corrects its rendering behavior
        // (https://github.com/swiftlang/swift-docc/issues/1611)
        if isMainSymbolGraph || graph.module.bystanders != nil {
```

QuietMisdreavus:
I'm fine to land this in principle, but i want to make sure we can make this function right. I opened https://github.com/swiftlang/swift-docc/issues/1611 in Swift-DocC to investigate how we can correctly load cross-import overlay module data from extensions, even if SymbolKit reports the symbol as belonging to the extended module. Going from the linked radar, it looks like more than the rendered module list is being affected by this change, so i think that a more thorough fix should be investigated.

I'm leaving my review as "changes requested" so that you can address some outstanding comments:

1. Please update the copyright year in `GraphCollectorTests.swift`, like @snprajwal mentioned.
2. Please change the test back to declaring separate graphs and "is extension" flags. Making the variables mutable makes the test more confusing, especially since the test symbol graph file name makes it look like there's a copy/paste error in the test.
3. Please link https://github.com/swiftlang/swift-docc/issues/1611 in the code somewhere in a FIXME comment so that we can refactor it back in the future. I want the `moduleNameFor` function to correctly sort extensions into their appropriate module name, regardless of whether they come from a regular module or a cross-import overlay, so i don't want that work to fall through the cracks. (To be fair, https://github.com/swiftlang/swift-docc-symbolkit/pull/102 should have refactored the comments around when it changed the conditional, since it made the comment directly below the condition nonsensical.)

Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift
snprajwal: Not in isolation, but I think it would be good to document this PR's changes better in general, because they seem confusing, and we don't want another ping-pong revert run due to the changes not being documented.

sofiaromorales: Resolved here https://github.com/swiftlang/swift-docc-symbolkit/pull/114/changes/03a49370bb625b6f72a944ffe89bf5bc18e5ca54

Tests/SymbolKitTests/UnifiedGraph/GraphCollectorTests.swift
sofiaromorales: Resolved here https://github.com/swiftlang/swift-docc-symbolkit/pull/114/changes/03a49370bb625b6f72a944ffe89bf5bc18e5ca54

sofiaromorales:
All the comments have been resolved as part of the latest commit.

>  Please update the copyright year in GraphCollectorTests.swift, like @snprajwal mentioned.

✅

> Please change the test back to declaring separate graphs and "is extension" flags. 

I reimplemented the test so it uses a different property instead of mutating an existing one +  renamed them to `overlayAtBName`/`overlayAtBIsMain`, since for `_A_B@B.symbols.json` the returned value is the declaring module (`A`) and not the extended one. Let me know if this is not what you are expecting.

> Please link https://github.com/swiftlang/swift-docc/issues/1611 in the code somewhere in a FIXME comment so that we can refactor it back in the future. 

Added the comment that you proposed.

snprajwal (commented):
Thanks for addressing the suggestions, LGTM. I'll let Vera greenlight the changes.

sofiaromorales:
@swift-ci please test

d-ronnqvist:
> @swift-ci please smoke test

AFAIK we don't have any smoke tests for this repo. I wonder if Joe was referring to doing a cross-repo smoke test (which I _think_ needs to be ran from the main Swift repo)

QuietMisdreavus (approved):
Thanks for putting up with my back-and-forth and investigations. ❤️ This is good to go now.

QuietMisdreavus:
@swift-ci Please test

PR #116, opened 2026-08-05
Clarify the documentation for the `prose` symbol name

timdream:
Bug/issue #, if applicable: None

## Summary

Swift-DocC now consumes this field: swiftlang/swift-docc#1557 uses `names.prose` as the link text for inline symbol links, falling back to `names.title` when it's absent.

Describe that inline-link use, and state the fallback to `title` in the OpenAPI schema.

## Dependencies

None

## Testing

N/A

## Checklist

- [ ] Added tests - N/A
- [ ] Ran the `./bin/test` script and it succeeded - N/A

### DIFF
--- a/Sources/SymbolKit/SymbolGraph/Symbol/Names.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/Names.swift
@@ -31,7 +31,7 @@ extension SymbolGraph.Symbol {
         public var subHeading: [DeclarationFragments.Fragment]?
 
         /**
-         A name to use in documentation prose or inline link titles.
+         The name of a symbol when referred to inline in documentation prose, such as the text of an inline link to the symbol.
 
          > Note: If undefined, use the `title`.
          */
--- a/Sources/SymbolKit/SymbolGraph/Symbol/Names.swift
+++ b/Sources/SymbolKit/SymbolGraph/Symbol/Names.swift
@@ -31,7 +31,7 @@ extension SymbolGraph.Symbol {
         public var subHeading: [DeclarationFragments.Fragment]?
 
         /**
-         The name of a symbol when referred to inline in documentation prose, such as the text of an inline link to the symbol.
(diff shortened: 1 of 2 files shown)


Sources/SymbolKit/SymbolGraph/Symbol/Names.swift
```diff
@@ -31,7 +31,7 @@ extension SymbolGraph.Symbol {
         public var subHeading: [DeclarationFragments.Fragment]?
 
         /**
-         A name to use in documentation prose or inline link titles.
+         The name of a symbol when referred to inline in documentation prose, such as the text of an inline link to the symbol.
```
d-ronnqvist: I find most of the added words to not provide new detail while also making this documentation inconsistently phrased compared to the other "names" above.

How do you feel about this rephrasing?

```suggestion
         An abbreviated form of the symbol's title that's preferred for use in documentation prose and inline link text.
```

openapi.yaml
```diff
@@ -263,7 +263,7 @@ components:
               items:
                 $ref: "#/components/schemas/DeclarationFragment"
             prose:
-              description: The name of a symbol when shown inline with other prose.
+              description: The name of a symbol when referred to inline in documentation prose, such as the text of an inline link to the symbol. If absent, `title` is used instead.
```
d-ronnqvist, line 266: I don't think it's correct for this specification to say that "`title` **is** used instead". AFAIK there's no logic in SymbolKit that uses the title as a fallback for the prose field. Rather, it's the callers responsibility to check the prose field and fall back to the title field if the prose field is not set.

timdream:
```suggestion
              description: An abbreviated form of the symbol's title that's preferred for use in documentation prose and inline link text. If absent, use `title`.
```

@d-ronnqvist Can SymbolKit document an instruction to the implementation here? The same thing is said in `Name.swift:36`.

